PR #1, opened 2019-08-16
Create medianSortedArray.java

mahima95:


super30admin:
Good job with intersect arrays problem. The median problem is brute force. 
Also attempt h index problem

PR #54, opened 2020-03-22
Completed Binary-Search-4. Please check.

billbz99:


H-Index-II.java
```diff
@@ -0,0 +1,28 @@
+// Time Complexity: O(N) best case, O(logN) worst case
```
rajatkumar01: Please review complexity. its opposite.

Average case (log(n) and worst case o(n)

rajatkumar01:
Good work Bill!!

PR #163, opened 2020-12-21
completed please review problem1

Mohit0993:


problem1.py
```diff
@@ -0,0 +1,36 @@
+#Time Complexity : ?? i am confused what will be the time complexity
```
samarthjj, line 1: Time here would be max of n log(n), m log(m), n log(m)

samarthjj:
Good job Mohit, Please review the inline comments.
Please also complete the other problem as well.

samarthjj:
Good job Mohit, Please review the inline comments.
Please also complete the other problem as well.

PR #215, opened 2021-02-15
Both problems done after the class

nilesh-525:


khushal1198 (commented):
Good job Nilesh!

khushal1198:
Good job Nilesh!

PR #490, opened 2022-08-13
Completed Intersection2Arrays and Median 2SortedArrys-Please review my Submission

Butulparveen:


intersection2Arrays2PointerApproach.java
```diff
@@ -0,0 +1,38 @@
+//TC for sorted = O(mlogm)+O(nlogn)
+//SC =O(1)
```
daredevilteja, line 2: SC is O(log k ) to O(k) where k is max(m,n)  
here space used by result should also be considered

intersection2ArraysBinarySearch.java
```diff
@@ -0,0 +1,40 @@
+//TC for unsorted = O(mlogm)+O(nlogn)+O(mlogn) and for sorted = O(mlogn)
+//SC =O(1)
```
daredevilteja, line 2: same as above

daredevilteja:
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #508, opened 2022-09-12
Completed. Please Review.

Thirunaa:


Intersections of Two Arrays II.py
```diff
@@ -0,0 +1,33 @@
+# TC : O(NlogN) - N is the length of the longest array
+# SC : O(1)
```
daredevilteja: SC is O(n) due to sorting

daredevilteja (changes requested):
Please mention if it ran successfully on leetcode or not, for every problem

daredevilteja:
Good Work! Keep it up

Intersections of Two Arrays II.py
Thirunaa: Please mention if it ran successfully on leetcode or not, for every problem - Sure
SC is O(n) due to sorting ? - Sorting happens in place. Right?

daredevilteja: To get TC as O( n log n) for sorting you have to use either merge sort or quick sort and for both the algorithms SC is not O(1), it would be in between O(log n) to O(n)

Thirunaa: Ohh got it, I thought Tim sort happens in place.

PR #551, opened 2022-12-19
Completed | Please review

Reethuch:


P1.py
```diff
@@ -0,0 +1,34 @@
+# TC : O(n log n)
+# SC : O(1)
```
daredevilteja, line 2: Recheck SC

daredevilteja:
Good Work! Keep it up.

PR #881, opened 2025-07-05
 Completed find median of two sorted array and intersection of two arrays

omkar04gaikwad:


super30admin:
The student's solution demonstrates a good understanding of the problem and provides two different approaches for finding the intersection of two arrays. The first method uses a hash map (Counter) to count occurrences, and the second method uses sorting and a two-pointer technique. Both methods are correctly implemented and handle edge cases such as empty arrays or no intersection. The time and space complexities are accurately noted for each method.

For the median of two sorted arrays problem, the student correctly implements a binary search approach on the smaller array, which is efficient and handles edge cases well. The solution correctly computes the median for both even and odd total lengths of the combined arrays. The time complexity is O(log(min(n, m))) and space complexity is O(1), which is optimal for this problem.

Strengths:
1. Clear and well-commented code.
2. Correct handling of edge cases.
3. Accurate time and space complexity analysis.
4. Efficient use of binary search for the median problem.

Areas for Improvement:
1. In the `intersection_linear` method, there is a typo where `result` is used instead of `res`, which would cause an error.
2. The method names could be more descriptive (e.g., `intersection_linear` could be `intersection_counter`).
3. For the median problem, the return value of -1 in the last line is unnecessary since the loop will always find a valid partition.
4. Adding more comments explaining the binary search logic for the median problem could enhance readability.

PR #891, opened 2025-07-28
solved median and intersection(used bs on just 1 of arrays

IshaaS:


super30admin:
- Strengths: The student has implemented a correct and efficient solution using binary search. The code is well-commented and follows good practices. The approach is innovative and shows a good understanding of the problem.
- Areas for Improvement: While the binary search approach is correct, the two-pointer approach after sorting might be simpler and easier to understand. The student could also consider mentioning the trade-offs between different approaches (hashmap, two-pointer, binary search) in the comments.

PR #892, opened 2025-08-06
Binary search 4 done

GuptaDivij:


super30admin:
Strengths:
- The student has demonstrated a strong understanding of the problems and has provided optimal solutions.
- The use of HashMap for frequency tracking in the first problem is efficient and handles duplicates well.
- The binary search approach for the second problem is well-implemented and correctly handles edge cases.

Areas for Improvement:
- For the first problem, while the solution is correct, it might be worth mentioning that if the arrays are already sorted, a two-pointer approach could be used for O(1) space complexity (though this is not a requirement).
- For the second problem, the student could add a brief explanation of why they check `if(n<m) return findMedianSortedArrays(nums2, nums1);` to ensure the binary search is performed on the smaller array, which optimizes the solution.

PR #898, opened 2025-08-11
Completed intersection of two arrays

harsh23raval:


super30admin:
Strengths:
- Good use of binary search to potentially improve time complexity.
- Clean code structure and good use of helper functions.
- Proper handling of integer overflow in binary search.

Areas for Improvement:
- The solution needs to correctly handle duplicate elements in both arrays. Currently, it may miss some intersections or include extra elements.
- The space complexity claim is incorrect. The result list requires O(min(m,n)) space, which should be acknowledged.
- Consider using a two-pointer approach after sorting, which would be more straightforward and efficient for this problem (O(m log m + n log n) time and O(1) space if output storage is not considered).

PR #899, opened 2025-08-15
Binary-Search-4

thakuryash060:


super30admin:
Strengths:
- Excellent use of binary search to optimize the intersection problem, significantly improving time complexity.
- Proper handling of edge cases in the median of two sorted arrays problem.
- Clean and readable code with good variable naming.

Areas for Improvement:
- In Problem1.java, consider adding a comment explaining why you swap the arrays when nums1 is larger than nums2 (to optimize space and time).
- In Problem2.java, the return value of 45444.98 in case of no solution is arbitrary and could be confusing. Consider throwing an exception or returning a more meaningful value if no median is found (though logically it should always find a median for valid inputs).

PR #907, opened 2025-09-24
added solution for BinarySearch-4

satvikapadakanti:


super30admin:
Strengths:
- The student has attempted both problems and shows understanding of basic concepts.
- The `median.java` solution is well-implemented with proper binary search approach.

Areas for Improvement:
1. For the intersection problem:
   - The output array size should be dynamic, not fixed at 3.
   - Need to handle duplicates properly by removing matched elements from nums2.
   - Consider using more efficient approaches like sorting and two-pointer or hash maps.
2. For both solutions:
   - Add comments to explain the logic, especially for complex parts like the median calculation.
   - Use more descriptive variable names (e.g., 'partition' instead of 'par1').
   - Handle edge cases explicitly (empty arrays, no intersection, etc.).

PR #910, opened 2025-10-05
Binary-Search-4 completed

sathvikareddy18:


super30admin:
- For the `intersection` problem, you need to handle duplicates correctly. The current solution only adds unique elements to the result because of the `seen` set. Instead, you should decrement the count in `smap` and add the element to the result whenever it appears in both arrays, regardless of whether it has been seen before.
- The `median` solution is well-implemented and efficient. Good job on using binary search to achieve optimal time complexity.
- Consider adding comments to explain the logic, especially for complex parts like the binary search in the `median` solution.
- Ensure that the function names match the problem requirements. The `intersection` function should be named `intersect` if it is meant to match the reference solution.

PR #923, opened 2025-11-08
Solved

sethu95:


super30admin:
- Problem1.java: Your solution is correct but more complex than necessary. Instead of storing indices in HashMaps, consider using a frequency map to count occurrences of each element. This would simplify the code and improve space efficiency. For example, you could use a HashMap to count frequencies in nums2, then iterate through nums1 and decrement counts when matches are found.
- Problem2.java: Excellent job! Your solution is efficient, clean, and follows best practices. The binary search approach is well-implemented and handles edge cases correctly.
- General: Your code is well-structured and readable. For future problems, consider simpler approaches when possible to improve both time and space complexity.

PR #928, opened 2025-11-20
Median of 2 arrays & intersection of 2 arrays

ashishbadhe97:


super30admin:
Strengths:
- The student has provided multiple approaches to solve the problem, demonstrating a good understanding of different techniques.
- The code is well-commented and easy to follow.
- The solutions are efficient and optimized for both time and space complexity.

Areas for Improvement:
- The binary search approach for the intersection problem could be further optimized by using two pointers instead of binary search, which would reduce the time complexity to O(m + n) without the need for sorting.
- The comments could be more concise and focused on the key points to avoid clutter.

PR #933, opened 2026-02-08
BS-4 Solutions

dbkuppagiri:


super30admin:
Your solution for "Intersection of Arrays" is excellent. It is efficient, correct, and well-written. Here are some strengths and minor suggestions:

Strengths:
- You correctly identified that building a frequency map for the smaller array minimizes space usage.
- The code is clean and easy to understand.
- You handle the frequency counts correctly by decrementing and removing when count reaches zero.
- The time and space complexity are optimal.

Areas for improvement:
- Although the code is correct, you might consider using a plain object instead of a Map if the numbers are non-negative integers (as per constraints). However, Map is perfectly acceptable and may be preferred for clarity.
- The problem does not require the output to be in any particular order, so your solution is fine. However, if the problem required the output to be in the same order as in nums2, you would need to adjust. But that is not the case here.
- You included a solution for a different problem (findMedianSortedArrays) in the same file. Make sure to only submit the solution for the problem you are solving.

Overall, great job!

super30admin:
Your solution for the "Intersection of Arrays" problem is excellent. It is efficient, with O(m + n) time complexity and O(min(m, n)) space complexity, which is optimal. The code is clean and well-commented. 

One minor suggestion: You could consider using a plain object instead of a Map for the frequency count if the numbers are non-negative integers (as per constraints), but using a Map is perfectly acceptable and more general.

Note that you included a solution for a different problem ("Median of Two Sorted Arrays") in your submission. Make sure to only include the solution for the problem you are solving to avoid confusion.

Overall, great job on the intersect.js solution!

PR #934, opened 2026-02-09
Complete Binary search 4

paridhimalviya:
Intersection of Two Arrays II (https://leetcode.com/problems/intersection-of-two-arrays-ii/)

Median of Two Sorted Arrays (https://leetcode.com/problems/median-of-two-sorted-arrays)


super30admin:
Strengths:
- You provided multiple approaches, showing a good understanding of different techniques.
- The two-pointer solution is correctly implemented and efficient.
- The binary search approach is well-thought-out for the case when one array is large.

Areas for improvement:
- In the hash map solution, you should not recursively call the function without returning the result. Instead, you should swap the arrays if needed and proceed. For example:
  if nums2.count < nums1.count { return intersectUsingHashMap(nums2, nums1) }
  This way, you ensure you always process the smaller array first.
- Similarly, in the binary search solution, avoid recursive calls without return. Use the same swapping technique.
- The binary search function should be adjusted to handle the range correctly. Currently, it searches for the first occurrence, which is good, but note that the array is sorted, so we can use this to avoid duplicates.
- Also, in the hash map solution, you need to check the count before decrementing. You should only add to the result if the count is positive.

Example fix for hash map:
```
func intersectUsingHashMap(_ nums1: [Int], _ nums2: [Int]) -> [Int] {
    if nums2.count < nums1.count {
        return intersectUsingHashMap(nums2, nums1)
    }
    var hashMap = [Int: Int]()
    for num in nums1 {
        hashMap[num, default: 0] += 1
    }
    var result = [Int]()
    for num in nums2 {
        if let count = hashMap[num], count > 0 {
            result.append(num)
            hashMap[num] = count - 1
        }
    }
    return result
}
```

- For the binary search solution, ensure that you update the low and high correctly. The current implementation is good, but note that the binary search should be done on the larger array? Actually, in your code, you are sorting both and then iterating over the smaller one and searching in the larger one. But you have a condition to call recursively if nums2 is smaller, which is not returned. So fix that.

Overall, the ideas are correct, but the implementation has bugs that need to be fixed.

super30admin:
Strengths:
- You provided multiple approaches, showing a good understanding of different techniques.
- The code is well-commented and structured.
- You considered edge cases like array sizes.

Areas for Improvement:
1. The hash map solution has a critical error: when nums2 is smaller, you call `intersectUsingHashMap(nums2, nums1)` recursively but do not return the result. This means the function will continue with the original arrays and produce incorrect output. You should return the result of the recursive call immediately.
   Correction: 
   ```swift
   if nums2.count < nums1.count {
       return intersectUsingHashMap(nums2, nums1)
   }
   ```
2. In the hash map solution, you should decrement the count only when the element exists and then check if the count becomes zero to remove it. However, your current code decrements the count and then checks for zero, which is correct. But note: if the count is nil, you should not decrement. Your code already handles that.
3. The binary search solution might have an issue in the `binarySearch` function. The condition `if (mid == low || arr[mid-1] < arr[mid])` to find the first occurrence is correct, but you need to ensure that `mid-1` is within bounds. Also, when `low` and `high` are passed, you should use them correctly.
4. You included a solution for "Median of Two Sorted Arrays" in the same file, which is not related to the intersection problem. It's better to separate solutions for different problems into different files.
5. For the two-pointer and binary search solutions, you have unused variables (hashMap) which should be removed.
6. In the binary search solution, you are sorting both arrays, which is necessary. However, you should note that the time complexity is dominated by sorting.
7. Consider testing your solutions with example cases to ensure correctness.

Overall, you have a good grasp of the problem and multiple approaches, but the implementation needs to be corrected for the hash map solution and tested thoroughly.

PR #935, opened 2026-02-10
My_sols

SrinivasaSupreeth:
leetcode_4:
Binary search on the smaller array

Partition both arrays into left and right halves

Ensure left max ≤ right min

Compute median based on total length

leetcode_350:
Count elements of the first array using a hashmap

Traverse the second array and check availability

Add common elements while decreasing count

Ensures correct handling of duplicates

### DIFF
--- a/leetcode_350.py
+++ b/leetcode_350.py
@@ -1,0 +1,11 @@
+class Solution:
+    def intersect(self, nums1: List[int], nums2: List[int]) -> List[int]:
+        freq = Counter(nums1)
+        result = []
+
+        for num in nums2:
+            if freq[num] > 0:
+                result.append(num)
+                freq[num] -= 1
+
+        return result
--- a/leetcode_4.py
+++ b/leetcode_4.py
@@ -1,0 +1,29 @@
+class Solution:
+    def findMedianSortedArrays(self, nums1: List[int], nums2: List[int]) -> float:
+        if len(nums1) > len(nums2):
+            nums1, nums2 = nums2, nums1
+
+        n, m = len(nums1), len(nums2)
+        left, right = 0, n
+
+        while left <= right:
+            cut1 = (left + right) // 2
+            cut2 = (n + m + 1) // 2 - cut1
+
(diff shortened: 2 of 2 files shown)


super30admin:
Your solution for the "Intersection of Arrays" problem is excellent. Here are some strengths and areas for improvement:

Strengths:
- Correctness: Your solution correctly handles the requirement of including each element as many times as it appears in both arrays.
- Efficiency: The use of Counter from collections makes the solution efficient with O(m + n) time complexity, which is optimal for this problem.
- Readability: The code is clean, concise, and easy to understand.

Areas for Improvement:
- There are no major issues with the code. However, it's good practice to consider edge cases explicitly, though your solution already handles them well (e.g., empty arrays).

super30admin:
Your solution for the "Intersection of Arrays" problem is excellent. You used a frequency counter which is efficient and straightforward. The code is clean and easy to understand. Well done!

One minor point: in Python, the Counter class from collections is a convenient tool for counting frequencies. However, it's good to be aware that in an interview setting, if you cannot use built-in libraries, you might need to implement a frequency dictionary manually. But for Leetcode, using Counter is perfectly acceptable.

Also, note that the problem has follow-up questions. Your solution already addresses some of them:
- If the arrays are sorted, we could use two pointers to avoid extra space. But your solution is efficient even for unsorted arrays.
- If nums1's size is small compared to nums2, your solution is efficient because you build the frequency map for the smaller array (if you choose to count the smaller array). Actually, in your code, you are counting nums1. So if nums1 is smaller, that is efficient. But if nums2 is smaller, it might be better to count nums2 instead. However, your solution is symmetric in the sense that it doesn't matter which one is smaller because the frequency map is built for nums1. To optimize for the case when one array is much smaller, you can choose to build the frequency map for the smaller array. You can add a check at the beginning to swap if nums1 is larger than nums2? But the problem doesn't require that, and your solution is efficient enough.

For the follow-up about if nums2 is stored on disk and memory is limited, your solution only requires the frequency map of nums1 to be in memory. Then you can stream nums2 from disk one element at a time and update the result. So it is memory efficient.

Overall, your solution is correct and efficient.

PR #937, opened 2026-02-16
completed binary search 4

kkhevna9252:


super30admin:
Your solution for the Intersection of Arrays problem is well-thought-out and efficient. You used sorting and binary search to achieve a time complexity of O(n log n + m log m + n log m), which is better than the brute-force approach. The code is readable and structured.

Strengths:
- You correctly identified that by sorting the arrays and using binary search, you can reduce the time complexity.
- The binary search function is implemented to find the first occurrence of the target, which is crucial for handling duplicates correctly.
- You made a good decision to always make nums1 the smaller array to minimize the number of binary searches.

Areas for improvement:
- There is a minor typo in the binarySearch function parameters: you wrote `newlow` and `newhigh` in the function signature, but in the function body you used `low` and `high` correctly. Actually, upon closer look, the parameters are named `newlow` and `newhigh`, and you assign them to `low` and `high` variables. This is correct. So no issue.
- You might consider using the two-pointer technique if the arrays are sorted, which would be more efficient (O(n+m)) and simpler. However, since the problem does not assume sorted arrays, your approach is valid.

Note: You included a solution for a different problem (Median of Two Sorted Arrays) in the same file. It is better to separate solutions for different problems into different files.

super30admin:
Strengths:
- You have chosen an efficient approach by sorting and using binary search, which has better time complexity than the reference solution.
- You considered the size of the arrays and process the smaller array first to minimize the number of binary searches.
- The code is structured with a helper function for binary search.

Areas for Improvement:
1. The binary search function for finding the first occurrence needs to be tested thoroughly. The current condition `if (mid == low || nums[mid - 1] != nums[mid])` may not be correct. For instance, when `mid == low`, it returns `mid` without checking if there are duplicates to the left (which there aren't since it's the first element). This is correct. However, if `nums[mid] == target` and `nums[mid-1] == target`, we set `high = mid - 1` to search left. This is correct. But note that the function requires the array to be sorted in non-decreasing order. It should work, but it's always good to test with edge cases (e.g., all duplicates).

2. In the main function, after finding an element, you set `low = bsIndx + 1`. This is correct to avoid reusing the same element. However, if the binary search returns -1 (not found), the `low` pointer remains unchanged. This is correct because we want to continue searching from the same starting point for the next number? Actually, no: since the arrays are sorted, if an element in `nums1` is not found, the next element might be larger, so we should not change `low`. But wait: if the current element is not found, it means that all elements in `nums2` from `low` to `high` are either too small or too large? Actually, since the array is sorted, if we don't find an element, we should not advance `low` because the next element in `nums1` might be larger and we should search from the same `low`? However, if the current element is not found, it might be because the element is not present in `nums2` from `low` to `high`. But note that the binary search is done over the entire range [low, high] and if it returns -1, it means the element is not in that range. Then for the next element in `nums1` (which is sorted), it might be larger, so we should not reduce the search range. Actually, we should not change `low` if the element is not found. So the code is correct in that regard.

3. However, there is a potential issue: if the same number appears multiple times in `nums1` and we are skipping parts of `nums2` by moving `low`, we might miss some duplicates. For example, consider `nums1 = [2,2]` and `nums2 = [2,2]`. After processing the first 2, we find it at index 0 (first occurrence) and set `low = 1`. Then for the next 2, we binary search in `nums2` from index 1 to end. The binary search for 2 in [2] (starting at index1) will return index1 (if implemented correctly). So it should work.

4. But the binary search function might not return the first occurrence in the segment [low, high]. It returns the first occurrence in the entire segment? Actually, it does because it searches for the first occurrence by moving left when it finds a duplicate. So it should be correct.

5. However, let's test the binary search function with a simple case: 
   nums = [2,2], target=2, low=0, high=1.
   low=0, high=1 -> mid=0. Check: nums[0]==2 -> true. Then check: mid==0 (which is low) -> true, so return 0. Correct.

   Now, if we have nums = [1,2,2], target=2, low=0, high=2.
   First mid=1: nums[1]==2 -> true. Then check: mid==0? no. Then check nums[0]!=2 -> true (since 1 !=2), so return 1. But the first occurrence of 2 is at index1, which is correct.

   Now, if we have nums = [2,2,2], target=2, low=0, high=2.
   First mid=1: nums[1]==2 -> true. Then check: mid!=0 and nums[0]==2, which is equal to nums[1], so we set high=0. Then we search from low=0 to high=0: mid=0, which is equal to target and mid==low, so return 0. This is the first occurrence. Correct.

   So the binary search function seems correct.

6. But note: the binary search function might not work if the target is not present. It returns -1 correctly.

7. Another point: the solution uses `Arrays.sort(nums1)` and `Arrays.sort(nums2)`. In Kotlin, you need to import `java.util.Arrays` or use `nums1.sort()` which sorts in place. However, the code does not show the import statement. It might cause a compilation error. You should either import or use the Kotlin built-in sort.

8. In Kotlin, you can use `nums1.sort()` instead of `Arrays.sort(nums1)`, which is more idiomatic.

9. The solution converts the result list to an array at the end. This is acceptable.

10. The solution handles the case when `nums1` is larger by swapping the arrays. This is good.

11. However, the solution does not consider the follow-up questions explicitly. For the follow-up:
    - If the arrays are already sorted, then we can skip the sorting step and directly use two pointers to traverse both arrays, which would be O(n+m). This would be more efficient.
    - If nums1's size is small compared to nums2, then the current approach is good because we are processing the smaller array.
    - If elements of nums2 are stored on disk, and memory is limited, we cannot load all elements into memory. In that case, we can sort nums1 (which is small) and then read chunks of nums2 from disk and perform binary search. Alternatively, we can use a hash map for nums1 (if it fits in memory) and then stream nums2 from disk.

12. The current solution is efficient for the constraints, but for the follow-up scenarios, there might be

PR #939, opened 2026-03-21
Binary Search 4

shinjaneegupta:


super30admin:
###  Intersection of Arrays  (Intersection.py)
Your solution is well-optimized and handles the problem efficiently. You correctly identified that using a frequency map for the smaller array reduces space complexity. The code is clear and concise. However, consider adding comments to explain the choice of using the smaller array for the frequency map, which might help readers understand the optimization. Also, note that the problem constraints are small (up to 1000 elements), so even the reference solution would work, but your approach is superior for larger inputs. Keep up the good work!

VERDICT: PASS

---

### Median of Arrays (Median.py)
Your solution is excellent and meets the problem requirements efficiently. Here are some strengths and areas for minor improvement:

Strengths:
- You achieved the optimal time complexity of O(log(min(m, n))) and constant space complexity.
- The code is clean, well-commented, and easy to understand.
- You correctly handle the case where one array is larger than the other by swapping them to binary search on the smaller array.
- The variables are named clearly (L1, R1, L2, R2) which makes the partition logic easy to follow.

Areas for improvement:
- The condition `if L1 <= R2 and L2 <= R1` is correct, but note that when the total length is odd, the left partition should have one more element than the right. Your code handles this by setting `partY = (n1 + n2) // 2 - partX`, which ensures the left partition has one more element when the total is odd. This is correct, but it's worth noting that the formula for `partY` might be slightly confusing. You might add a comment explaining that `(n1+n2)//2` is the number of elements in the left partition for even total length, and for odd total, the left partition has one more, but the formula still works because `(n1+n2)//2` truncates towards zero, so when total is odd, `(n1+n2)//2` is actually `(total-1)//2`, which is the size of the left partition minus one? Actually, in your code, for odd total, the left partition should have `(n1+n2+1)//2` elements. However, your current formula `partY = (n1 + n2) // 2 - partX` might not be correct for odd total. Let me check: 
  For example, if n1=1, n2=0, total=1 (odd). Then partX should be 1? Because the left partition should have 1 element. But (1+0)//2 = 0, so partY = 0 - partX = -1? This is incorrect.

Wait, there is a critical error in the calculation of `partY`. The formula should be `partY = (n1 + n2 + 1) // 2 - partX` to ensure the left partition has the correct number of elements (one more when total is odd). Your current code uses `(n1+n2)//2 - partX`, which is incorrect for odd total lengths.

For instance, consider nums1 = [1], nums2 = []. The total length is 1 (odd), so the median should be 1.0. With your code:
  n1=1, n2=0 -> since n1>n2 is false? Actually, n1=1, n2=0 -> n1>n2 is true? Wait, your code first checks if n1>n2, so it will swap: now nums1 becomes [] and nums2 becomes [1]. Then n1=0, n2=1.
  low=0, high=0 -> partX = 0.
  partY = (0+1)//2 - 0 = 0.
  Then L1 = -inf (since partX=0), R1 = inf (since partX=0 and n1=0? Actually, for n1=0, partX=0 is valid, but when partX==n1, R1=inf.
  L2 = -inf (partY=0), R2 = nums2[0]=1.
  Check conditions: L1<=R2 (-inf<=1) and L2<=R1 (-inf<=inf) -> true.
  Then check total length: (0+1)=1 is odd -> return min(R1, R2)=min(inf,1)=1.0 -> correct.

But what if we have two non-empty arrays with odd total? For example: nums1=[1], nums2=[2,3]. Total=3 (odd). Since n1=1, n2=2 -> n1<=n2, so no swap.
  low=0, high=1.
  First mid: partX = (0+1)//2 = 0.
  partY = (1+2)//2 - 0 = 3//2 -0 = 1.
  Then:
    L1 = -inf (partX=0), R1 = nums1[0]=1.
    L2 = nums2[0]=2 (since partY=1, so index=1-1=0), R2 = nums2[1]=3.
  Check: L1<=R2 (-inf<=3) and L2<=R1? -> 2<=1? false. So we move low to partX+1=1.
  Now partX=1.
  partY = 3//2 -1 = 1-1=0.
  Then:
    L1 = nums1[0]=1 (partX=1, so index=0), R1 = inf (partX=1==n1=1? yes -> inf)
    L2 = -inf (partY=0), R2 = nums2[0]=2.
  Check: L1<=R2 (1<=2) and L2<=R1 (-inf<=inf) -> true.
  Total is odd -> return min(R1,R2)=min(inf,2)=2.0. But the correct median should be 2.0? Actually, the merged array is [1,2,3] -> median=2.0. So it works.

Wait, but why? The formula for partY is (n1+n2)//2 - partX. For total=3, (n1+n2)//2=1. So partY=1 - partX. When partX=0, partY=1; when partX=1, partY=0. This ensures that the left partition has partX + partY = 0+1=1 or 1+0=1 elements? But for odd total, the left partition should have (total+1)//2 = 2 elements? Actually, no: in the binary search method for median, we partition the arrays such that the left part has (n1+n2+1)//2 elements. So for total=3, left should have 2 elements. But in your code, when partX=1 and partY=0, the left part has partX (elements from nums1) + partY (from nums2) = 1+0=1 element? This is not enough.

Wait, there is a mistake in the student's code. The correct formula for partY should be `partY = (n1 + n2 + 1) // 2 - partX` to ensure the left partition has (n1+n2+1)//2 elements. The student's code uses `(n1+n2)//2 - partX`, which is incorrect.

Let me test with the same example: nums1=[1], nums2=[2,3] (total=3). The left partition should have 2 elements. So if partX=1 (meaning we take 1 element from nums1), then partY should be 2-1=1? So partY=1. Then:
  L1 = nums1[0]=1
  R1 = inf
  L2 = nums2[0]=2
  R2 = nums2[1]=3
  Then the left partition has [1,2] and right has [3]. Then max(L1,L2)=max(1,2)=2, min(R1,R2)=min(inf,3)=3. For odd total, median is min(R1,R2)=3? which is wrong.

Actually, the correct partition for this case should be partX=0? Then partY=2? But partY cannot be 2 because n2=2, so partY can be at most 2 (if we take all elements). So partX=0, partY

VERDICT: NEEDS_IMPROVEMENT

PR #942, opened 2026-03-25
Complete Binary-Search-4

PrakarshKamal:


super30admin:
###  Intersection of Arrays  (IntersectionOfArrays.java)
Your solution is excellent and demonstrates a good understanding of the problem and efficient use of data structures. Here are a few points for improvement:
1. The `map.remove(num, 0)` method is correct, but it is called for every element in nums2 that is in the map. This is not inefficient per se, but you could simply check if the frequency becomes zero and then remove the key to avoid multiple calls. However, the current approach is acceptable.
2. You might consider using a counter to preallocate the result array to avoid the overhead of ArrayList, but this is optional and not critical.
3. The comment "O(n+m) time, O(min(n,m)) space" is accurate, but you could clarify that the space is for the hashmap and the result list.
4. The code is clean and readable. Well done!

Overall, your solution is optimal and handles the problem correctly.

VERDICT: PASS

---

### Median of Arrays (MedianOfSortedArrays.java)
Note: The student has provided two solutions. The first one is O(m+n) and the second one is O(log(min(m,n))). The problem requires O(log(m+n)) which the second solution satisfies. However, we need to evaluate both solutions for correctness and efficiency.

Let's evaluate each solution separately.

First solution (Solution):
- Correctness: The solution correctly merges the two arrays and then finds the median. It should work correctly.
- Time Complexity: O(m+n) which is acceptable given the constraints (since m+n <= 2000) but does not meet the required O(log(m+n)).
- Space Complexity: O(m+n) for the merged list. Again, acceptable for the constraints but not optimal.
- Code Quality: The code is clear and readable. It uses a while loop to merge the arrays. However, using an ArrayList might be less efficient than using an array (due to boxing and dynamic resizing) but for the constraints it's acceptable.

Second solution (Solution2):
- Correctness: This solution uses the binary search method on the smaller array. The logic appears correct: it partitions both arrays such that the left half has the same number of elements as the right half (or one more in the odd case). It checks the boundary conditions (L1 <= R2 and L2 <= R1) to ensure the partition is correct. If not, it adjusts the binary search. This should correctly find the median.
- Time Complexity: O(log(min(m,n))) which is O(log(m+n)) as required.
- Space Complexity: O(1) which is optimal.
- Code Quality: The code is well-structured. It uses meaningful variable names. However, there is a potential issue: the condition for binary search might have an off-by-one error? Let's check: 
  The binary search runs from low=0 to high=n1 (which is the length of the smaller array). The partition index partX can range from 0 to n1. The partY is computed as (n1+n2)/2 - partX. Note that for the even and odd cases, the formula for partY should be adjusted? Actually, the total number of elements in the left partition should be (n1+n2+1)/2 to handle both even and odd? But in the code, it uses (n1+n2)/2 - partX. This might be incorrect.

Wait, let me verify the second solution with an example.

Example 1: nums1 = [1,3], nums2 = [2] -> total length = 3 (odd). The median should be at index 1 (0-indexed) in the merged array.

In Solution2:
  n1=2, n2=1 -> since n1>n2? Actually, the code checks if n1>n2 then swap. So here, n1=2 and n2=1 -> n1>n2, so it will call with swapped: nums2 and nums1. So the smaller array is nums2 (length=1). Then:
  n1=1, n2=2.
  low=0, high=1.
  partX = (0+1)/2 = 0? Then partY = (1+2)/2 - 0 = 3/2 = 1 (integer division) -> partY=1.
  Then:
    L1 = partX==0? -> MIN_VALUE
    R1 = partX==1? -> MAX_VALUE? Actually, partX=0 -> so R1 = nums1[0] = 1 (from the swapped array, which is nums2? Wait, after swapping: the first argument is the smaller array. So in the call with nums2 and nums1, the parameters are: 
        nums1 = [2] (the smaller) and nums2 = [1,3] (the larger).
  So:
    partX=0: 
        L1 = MIN_VALUE
        R1 = nums1[0] = 2
    partY=1:
        L2 = nums2[1-1] = nums2[0] = 1
        R2 = nums2[1] = 3? But wait, nums2 has length 2, so partY=1 is valid.

  Check condition: L1<=R2? (MIN_VALUE<=3 -> true) and L2<=R1? (1<=2 -> true). So we found the partition.
  Since total length=3 (odd), we return Math.min(R1,R2)=min(2,3)=2 -> correct.

  Now, what if partX=1?
    Let's try: partX=1: 
        L1 = nums1[0]=2
        R1 = MAX_VALUE
        partY = (1+2)/2 - 1 = 1 - 1 = 0
        L2 = MIN_VALUE
        R2 = nums2[0]=1
    Then condition: L1<=R2? 2<=1? false -> so we would adjust.

  So the binary search found partX=0 correctly.

  Now for even case: nums1=[1,2], nums2=[3,4] -> total length=4 (even). After swapping? n1=2, n2=2 -> no swap because n1 is not greater than n2. So we proceed with nums1 and nums2 as given.
  n1=2, n2=2.
  low=0, high=2.
  Let partX = 1 -> then partY = (2+2)/2 - 1 = 2-1=1.
  Then:
    L1 = nums1[0]=1
    R1 = nums1[1]=2
    L2 = nums2[0]=3
    R2 = nums2[1]=4
  Check: L1<=R2? 1<=4 -> true; L2<=R1? 3<=2 -> false -> so we need to adjust.

  Why false? Because the partition is not correct. We need to have the left partition having the first two elements? But here the left partition has [1] from nums1 and [3] from nums2? Actually, the left partition should have two elements: from nums1 we take partX=1 (so first 1 element) and from nums2 we take partY=1 (so first 1 element) -> total 2. The right partition has the rest. But the condition requires that the last element in the left partition of nums1 (1) is <= the first element in the right partition of nums2 (4) -> true. And the last element in the left partition of nums2 (3) is <= the first element in the right partition of nums1 (2) -> false. So we need to move partX to the right? Actually, the code says: else { low = partX+1 }.

  So next: low=2, high=2 -> partX=2. Then partY = (2+2)/2 - 2 = 2-2=0.
  Then:
    L1 = partX==2? -> if partX==n1? yes -> so L1 = nums1[1]=2? Wait, no: for partX=2, we are taking the first 2 elements from nums1. Then the left partition has all of nums1? Then for nums2, partY=0 -> no elements. So:
    L1 = nums1[1]=2
    R1 = MAX_VALUE
    L2 = MIN_VALUE
    R2 = nums2[0]=3
  Check: L1<=R2? 2<=3 -> true; L2<=R1? MIN_VALUE<=MAX_VALUE -> true. So we found the partition.
  Then since even: return (max(L1,L2) + min(R1,R2))/2 = (max(2, MIN_VALUE)=2 + min(MAX_VALUE,3)=3) /2 = (2+3)/2=2.5 -> correct.

  So the formula for partY: (n1+n2)/2 - partX. This works for both even and odd? Actually, for the odd case: (n1+n2) is odd -> (n1+n2)/2 is integer division? In Java, integer division truncates toward zero. So for total=3: (1+2)/2 = 1. Then partY=1 - partX. This gives the correct partition.

  However, note that the left partition should have (n1+n2+1)/2 elements to account for both even and odd? Actually, the common approach is to let the left partition have (n1+n2+1)/2 elements. Then the median is computed from the left partition. But in this code, the left partition has partX + partY = partX + ((n1+n2)/2 - partX) = (n1+n2)/2. For odd total: (n1+n2)/2 is integer division. For example, total=3: (3)/2=1. But the left partition should have 2 elements? Wait no: in the common binary search solution for median, we typically partition such that the left part has (m+n+1)/2 elements. This ensures that in the odd case, the median is the maximum of the left part. In the even case, the median is (max(left) + min(right))/2.

  But in this student's solution, the left partition has partX + partY = (n1+n2)/2 elements. For total=3: left partition has

VERDICT: NEEDS_IMPROVEMENT

PR #945, opened 2026-05-25
Intersection of Two Arrays II and Median of Two Sorted Arrays

anirudhv98:
Completed Binary-Search-4


super30admin:
###  Intersection of Arrays  (Problem1.cs)
**Strengths:**
- Good documentation and comments explaining the approach
- Clean code structure with a separate helper method
- Handles the case where nums2 is smaller by swapping arrays
- Good use of C# features like GetValueOrDefault

**Areas for Improvement:**
1. **Critical Bug**: The algorithm iterates over nums1 directly, which means if nums1 has duplicate values, it will only find one occurrence in nums2. For example, if nums1=[2,2] and nums2=[2,2,2], the result should be [2,2] but the algorithm would only return [2].

2. **Logic Fix**: Instead of iterating over nums1, you should iterate over the keys of the frequency map. For each unique value, find its first occurrence in nums2, then continue finding subsequent occurrences by updating the search start position.

3. **Complexity**: The binary search approach adds unnecessary complexity. A hash map approach (like the reference solution but using HashSet/Map instead of list) would be O(m+n) time and simpler to implement correctly.

4. **Testing**: Consider edge cases like:
   - nums1 = [1,1,2], nums2 = [1,2,2,2] (should return [1,2,2])
   - nums1 = [1,2,2], nums2 = [2,2] (should return [2,2])

VERDICT: NEEDS_IMPROVEMENT

---

### Median of Arrays (Problem2.cs)
**Strengths:**
- Excellent implementation of the binary search approach - this is the optimal solution for this problem
- Time complexity of O(log(min(m,n))) significantly outperforms the O(m+n) reference solution
- Space complexity of O(1) is optimal, compared to O(m+n) in the reference
- Good defensive coding: swapping arrays when nums1 > nums2 ensures binary search runs on the smaller array
- Clear comments explaining the algorithm logic
- Proper handling of edge cases (empty partitions) using sentinel values

**Areas for improvement:**
- The sentinel values (Int32.MinValue/MaxValue) could potentially cause overflow issues if actual array values are at the extremes. A cleaner approach would be to use nullable types or conditional checks.
- The final return -1 is unreachable in practice; consider throwing an exception or using a more meaningful return value for robustness.

**Minor suggestions:**
- Could add null checks for input arrays for production code
- The comment about "size of array + 1" partitions is slightly confusing - it should clarify that partitions can be from 0 to n1 (inclusive)

VERDICT: PASS

PR #946, opened 2026-05-28
Binary-Search-4 Assignment Completed

Hriday-A:


super30admin:
###  Intersection of Arrays  (intersection_of_2_arrays.java)
**Strengths:**
- Good initiative to sort arrays and use binary search (better than brute force)
- Clean code structure with proper separation of concerns
- Handles the case where nums1 is larger by swapping arrays
- Good variable naming and documentation

**Areas for Improvement:**
- **Critical Bug**: The binary search finds only the first occurrence of duplicates, causing incorrect results
- Consider using a HashMap to count frequencies instead of binary search
- Alternative: Use two-pointer approach after sorting (more efficient for sorted arrays)
- The current approach doesn't properly track how many times each element has been matched

**Suggested Fix:**
Use a HashMap to count occurrences in nums2, then iterate through nums1:
```java
Map<Integer, Integer> count = new HashMap<>();
for (int num : nums2) {
    count.put(num, count.getOrDefault(num, 0) + 1);
}
// Then decrement counts as you find matches
```

VERDICT: NEEDS_IMPROVEMENT

---

### Median of Arrays (median_of_2_sorted_arrays.java)
**Strengths:**
1. Excellent approach - using binary search instead of merging shows deep understanding of the problem
2. Handles all edge cases properly (empty arrays, single elements)
3. Space efficient with O(1) auxiliary space
4. Clean, readable code structure
5. Correctly swaps arrays when needed to ensure binary search is performed on smaller array

**Areas for Improvement:**
1. The use of Integer.MIN_VALUE and Integer.MAX_VALUE as sentinel values works but could be improved by using Long types or Double.POSITIVE_INFINITY/NEGATIVE_INFINITE for better handling of edge cases with extreme values
2. Could add a brief comment explaining the binary search partition logic for future maintainability
3. The final return -1 is unreachable but could be replaced with a more descriptive exception or comment

**Minor optimization:** The condition `low + (high - low)/2` is correct but could also be written as `(low + high) >>> 1` for unsigned right shift (though in Java this is less critical than in C++)

VERDICT: PASS

PR #951, opened 2026-07-02
Complete Binary Search 4

MrunaliVaidya98:


super30admin:
### Intersection of Two Arrays II (Problem2_ findMedianSortedArrays.java)
- **Major Issue**: You have submitted a solution for the wrong problem. The problem asks for "Intersection of Two Arrays II" but your code solves "Find Median Sorted Arrays".
- Your code for the median problem is actually well-written with proper edge cases, binary search logic, and correct handling of even/odd total lengths.
- If you intended to solve the intersection problem, you would need to use a HashMap to count frequencies or sort both arrays and use two pointers approach.
- Please ensure you are submitting solutions for the correct problem statement.

VERDICT: NEEDS_IMPROVEMENT

---

### Median of Two Sorted Arrays (Problem1_intersect.java)
- The code quality itself is good - well-structured, readable, and follows Java best practices
- The solution correctly solves the "Intersection of Two Arrays" problem
- However, this is not the problem that was assigned - the code does not compute the median of two sorted arrays
- To solve the actual problem, the student would need to implement a merge-based approach or binary search to find the median in O(log(m+n)) time

VERDICT: NEEDS_IMPROVEMENT

